PR #3, opened 2019-07-10
binary array and subarraySumEqualK completed- please review

PremalSamale:


ContiguousArray.java
```diff
@@ -0,0 +1,43 @@
...
+//Space complexity :o(n)Maximum size of the HashMap map will be n, if all the elements are either 1 or 0.
+//Did this code successfully run on Leetcode : yes
+//Any problem you faced while coding this :no
+
+public class ContiguousArray {
+	public static int findMaxLength(int[] nums) {
+	//declare hashmap with key= count and value = current index
+    Map<Integer, Integer> map = new HashMap<>();
+    //initalize map with key =0 and value =-1
+    map.put(0, -1);
```
ashmeetdeol19: Can you state the reason for doing so ?

SubarraySumEqualsK.java
```diff
@@ -0,0 +1,40 @@
...
+        //return count which is no of  subarray found with target sum as k
+        return count;
+    }    
+
+    public static void main(String[] args) {
+    	int[] nums = {1,1,1};
+    	int targetSum =2;
+    	int count = subarraySum(nums,targetSum);
+        System.out.println("number of subarrays with targetsum="+count);	
+    }
```
ashmeetdeol19: Well done !!

ContiguousArray.java
PremalSamale: count 0 means no. of occurrences of 1 and 0 ares equal and which indicate got subarray with equal 1 and 0.  so to increase count by 1 for 0 keeping value  -1. As length =end index -(start index) so -1 become +1.

PR #4, opened 2019-07-10
Please review problem 1 and problem 2

AkhilKrishnamurthy:


ashmeetdeol19:
Its good to see that you have tried brute force if you are not getting to solution.
Solution to problem 2 was great.
Well Done !!

PR #6, opened 2019-07-10
Completed Day 6 problem 1 and 3, please review

MahiteeS:


LongestPalindrome.java
```diff
@@ -0,0 +1,40 @@
+package day6;
+
+import java.util.*;
+//Time Complexity : O(n) where n is length of s
+// Space Complexity : O(1) [space required by count variable]
```
ashmeetdeol19: What about space by HashSet ?

ashmeetdeol19:
Well done with both of the Solutions.
Kindly Submit Problem 2 as well.

LongestPalindrome.java
MahiteeS: yes, O(n) for iterating over the array and storing elements in Hashset.

PR #11, opened 2019-07-10
Implemented sub-array, LargestArray, LongestPallindromeSubstring

PrachiJani13:
Please review


super30admin:
Very well Done in Problem_1.
For the other Two Problems instead of using Brute Force, Go for the optimal solution as we discussed.

PR #16, opened 2019-08-06
Completed Problem 1 and 3: Please review

muditabysani:


Problem_1.py
```diff
@@ -0,0 +1,25 @@
...
+			temp_sum = arr[i]
+			if temp_sum == k:
+				count += 1
+				continue
+			for j in range(i + 1, len(arr)):
+				temp_sum += arr[j]
+				if temp_sum == k:
+					count += 1
+					break
+				elif temp_sum > k:
```
kaurramanpreet: you cannot use break here because the array values can be negative as well

Problem_1_HashMap.py
```diff
@@ -0,0 +1,21 @@
+class Problem_1:
+	def findContinuousArray(self, arr, target):
+		runningSumDict = dict()
+		count = 0
+		runningSumDict[0] = 1
+		runningSum = 0
+		for i in arr:
+			runningSum += i
+			compliment = runningSum - target
```
kaurramanpreet: it should be target-runningsum

Problem_1.py
muditabysani: But if the sum is equal to the target, I can break the loop right

Problem_2.py
```diff
@@ -0,0 +1,8 @@
+class Problem_2:
+	def EqualNumbersCount(self, arr):
+		pass
```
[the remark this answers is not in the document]
muditabysani: I didn't know how to implement that problem. So just added the function name and wrote pass. It will compile perfectly fine but does nothing

Problem_1_HashMap.py
muditabysani: It has to be runningSum - target right? Because the running sum keeps on increasing and it will be greater than target. And as far as I remember thats how the algorithm works. In the class, the examples always had the positive compliment to be checked in the hashmap

PR #18, opened 2019-08-06
hashing 2

pranshugoyal0111:


EqualZeroOne.java
```diff
@@ -0,0 +1,68 @@
+//Not sure if all test cases will pass. Gave it a try
+f//irst replaced all 0 with -1. Now we have to find subarrays which have sum =0
+//if sum =0 and it is not already present in the hashmap , we add that index as value  at which sum becomes 0 and sum as key. size in this case will be 0->index
+//if we again get sum =0 at new index and sum=0 is already present in map(let this be y). suppose old index at which sum=0 was x,
+//this means array from 0 to x has sum=0 and 0 to y also has sum which implies x+1 to y is also 0. Therefore we iterate entire arraylist at this key and form subarray combinations with old indexes already in hashmap and the new index
```
kaurramanpreet: dont you think instead of it to be zero for both indices it should be equal to each other such that their difference come out to be zero which implies the numbers(subarray) between these two indices is giving the sum 0

PR #20, opened 2019-08-06
Hashing 2 done - Please review

swaps95shah:
Grasped the 3 problems well. Please review the code. Thanks


problem3.py
```diff
@@ -0,0 +1,40 @@
+"""
+time: O(n)
+space:O(n)
+Leet: accepted 79.86 percentile
```
kaurramanpreet: what does this mean?

swaps95shah: Means that my code surpassed roughly 80% of the submitted python codes on Leetcode

kaurramanpreet: ohh that doesnt matter
it depends on the testcases 
we dont need to bother about that

PR #32, opened 2019-10-16
problems 1 and 2

vramanathan:


vramanathan:
added problem3 palindrome

PR #34, opened 2019-10-16
Completed 

pradeepikag:
Completed 16 and 17.
But a few test cases failed in 17 though, will go through it again.


sehaj-grover:
great 
just try implementing them again after the session 
If stuck just ping me with the problem

PR #36, opened 2019-10-16
Hashing-2

Rohanp1196:
Problem-1 Solution added. Brute force- Not optimal


Rohanp1196:
Will be adding other problems

PR #53, opened 2019-11-27
Adding solutions for Hashing-2 repo

amitkc1:


LongestPalindrome.java
```diff
@@ -0,0 +1,24 @@
+class LongestPalindrome {
+    public int longestPalindrome(String s) {
+        HashSet <Character> palindromeMap = new HashSet<>();
```
neha79091: Write complexities of the solution with your approach.

PR #54, opened 2019-11-27
Updating Hashing 2 changes

dasgprat:
Please review the PR with techniques learned from class today.

### DIFF
--- a/Problem2.java
+++ b/Problem2.java
@@ -1,0 +1,38 @@
+/**
+Leetcode Submitted : YES
+Time Complexity : O(N)
+Space Complexity : O(1)
+The approach is to keep running sum in Hashmap by calculating complement with target sum at each iteration.
+Once, the runsum is found increase its count
+**/
+
+
+class Solution {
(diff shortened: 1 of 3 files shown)


Problem2.java
```diff
@@ -0,0 +1,38 @@
+/**
+Leetcode Submitted : YES
+Time Complexity : O(N)
+Space Complexity : O(1)
+The approach is to keep running sum in Hashmap by calculating complement with target sum at each iteration.
+Once, the runsum is found increase its count
```
neha79091: Space complexity will not be constant.

Problem3.java
```diff
@@ -0,0 +1,37 @@
+
+/**
+Leet Code Submitted : YES
+Time Complexity : O(N)
+Space Complexity : O(1)
```
neha79091: Space complexity will not be constant.

dasgprat: Yeah I will fix it

Problem2.java
dasgprat: Yeah I will fix it

PR #59, opened 2019-12-02
Problem 1 and 3

Jasmine9003:
Kindly review @neha79091


Problem3.java
```diff
@@ -0,0 +1,34 @@
+// Time Complexity : O(N)
+// Space Complexity : O(N)
+// Did this code successfully run on Leetcode : yes
```
neha79091: Space complexity will be constant

PR #65, opened 2020-01-01
Please review

nikrasiya:


560_sub_array_sum_equals_k.py
```diff
@@ -0,0 +1,19 @@
+from collections import Counter
+
+
+class Solution:
+    def subarraySum(self, A, K):
+        count = Counter()
+        count[0] = 1
```
super30admin: What is the implementation of counter under the hood

nikrasiya: Counter is an implemented using dictionary. It's the same thing. It's just its a better DS to use in Python when you are counting something. It does error handling and some optimization.

super30admin: Ohh that is counter. Great to learn

PR #77, opened 2020-01-06
please review third problem 

Nsharma335:


rajatkumar01:
Good Job Neha. You have attempted all the problems very well

PR #81, opened 2020-02-01
Completed "Subarray Sum k", "Contiguous Subarray" and "Longest Palindrome" 

ashu20red:


Problem2.java
```diff
@@ -0,0 +1,43 @@
+/* Given a binary array, find the maximum length of a contiguous subarray with equal number of 0 and 1. */
+
+// Time Complexity : O(n) using hashmap
+// Space Complexity : O(1) in place
+// Did this code successfully run on Leetcode : Yes
```
neha79091: Space complexity will not be constant as we are using hashmap also

Problem2.java
```diff
@@ -0,0 +1,43 @@
...
+// Any problem you faced while coding this : No
```
neha79091: Space complexity will not be constant as you are using hashmap also

ashu20red: Thanks for your observation. I have updated space complexity.

neha79091: good

PR #96, opened 2020-02-16
completed 

xlsarath:
pls review and let me know if the entered time and space complexity are fine .

### DIFF
--- a/SubArraySumEqualsK.java
+++ b/SubArraySumEqualsK.java
@@ -1,0 +1,48 @@
+/**
+ * time complexity :  O(n)
+ * space complexity : O(1)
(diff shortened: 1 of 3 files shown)


SubArraySumEqualsK.java
```diff
@@ -0,0 +1,48 @@
+/**
+ * time complexity :  O(n)
+ * space complexity : O(1)
+ * leet code execution successfull
```
super30admin: Space is O(n) as we are using an auxiliary data structure hashmap

super30admin:
Space is O(n) as we are using an auxiliary data structure hashmap

PR #106, opened 2020-03-23
Completed Hashing-2: Please review 

KBangera:


Contiguous_Array.java
```diff
@@ -0,0 +1,24 @@
+// Time Complexity : O(n * sizeOfMap)
```
devashishnyati: Why n * sizeOfMap???

Longest_Palindrome.java
```diff
@@ -0,0 +1,34 @@
+// Time Complexity : O(n * sizeofmap)
```
devashishnyati: Here also, why sizeofmap? Putting in hashmap and getting from hashmap is O(1).

devashishnyati:
Good job Karthik!
Check the time complexities! Putting and getting from hashmap is O(1).

PR #135, opened 2020-04-30
Jay Nirmal PR Hashing 2

jaynirmal15:


SubarraySumEqualsK.java
```diff
@@ -0,0 +1,20 @@
+package s30Coding;
+
+import java.util.*;
+//Time Complexity :- O(n)
+//Space Complexity :- O(n)
+
```
sowmyadvn, line 6: Short explanation of the approach is good idea to refer later

sowmyadvn:
Solution is correct Jay. Try solving contiguous sub array and longest palindrome problems as well.

sowmyadvn:
Solution is correct Jay. Try solving contiguous sub array and longest palindrome problems as well.

PR #136, opened 2020-04-30
Sprint 1 of hashing 2

Akshay-Ravi:


longest_palindrome.cpp
```diff
@@ -0,0 +1,23 @@
+class Solution {
```
super30admin, line 1: Time and space here?

longest_palindrome.cpp
```diff
@@ -0,0 +1,23 @@
...
+public:
+    int longestPalindrome(string s) {
+        //We need to add the alphabets in a hashmap and map them to their counts
```
super30admin, line 4: Hashset

super30admin:
Good Job Akshay

longest_palindrome.cpp
Akshay-Ravi: Space is going to be constant cause maximum number of elements in the hashset will be the total number of alphabets. Time complexity will be O(string size) because we are going to traverse through the string. Please correct me if I'm wrong. @super30admin

super30admin: Time?

PR #149, opened 2020-06-01
Hashing-2: Added longest palindrome with wrong answer

karthikprabhuA:


SubArraySum.swift
```diff
@@ -0,0 +1,25 @@
+
+// Time Complexity : (O)n
```
chiragshahi, line 2: define n

SubArraySum.swift
```diff
@@ -0,0 +1,25 @@
...
+// Space Complexity : (O)1
+// Did this code successfully run on Leetcode : No. Wrong answer
```
chiragshahi, line 4: Make use of hashing DS
store in map -> number of occurences of a particular element
iter thru arr and if running sum-k is found in map, add to count the mapped value of running sum-k

LongestPalindrome.swift
```diff
@@ -0,0 +1,34 @@
+// Time Complexity : (O)n
+// Space Complexity : (O)1
+// Did this code successfully run on Leetcode : No. Wrong answer
```
chiragshahi, line 3: Please go thru this java code and try again.
public int longestPalindrome(String s) {
        if (s.length() < 2) return s.length(); 
        HashSet<Character> set = new HashSet<>();
        int ans = 0;
        for (char c : s.toCharArray()) {
            if (!set.contains(c)) set.add(c);
            else {
                ans += 2;
                set.remove(c);
            }
        }
        return set.isEmpty() ? ans : ans + 1;
    }

LongestPalindrome.swift
```diff
@@ -0,0 +1,34 @@
+// Time Complexity : (O)n
```
chiragshahi, line 1: define n

chiragshahi:
Nice try! Kindly refer to the inline comments for detailed approaches for both the problems 
Also try the 3rd problem

PR #154, opened 2020-06-02
Complete 2 problems from Hashing-2

nileshdev996:


src/LongestPalindrome.java
```diff
@@ -0,0 +1,54 @@
+import java.util.HashMap;
+import java.util.Map;
+
+// Time Complexity : O(n) n = char in the string
+// Space Complexity : O(1)
```
Lohith-Aswathappa, line 5: You are using hashmap in this code, consider that for the space complexity

Lohith-Aswathappa:
Good job, Nilesh!

PR #183, opened 2020-06-03
Please review

snigdharreddy:


AbhiniveshP:
Good job on both the questions. Please implement the other question as well.

AbhiniveshP:
Good job on both the questions. Please implement the other question as well.

PR #196, opened 2020-07-14
solutions to 3 problems___please review!

khgupta:


Problem3.py
```diff
@@ -0,0 +1,20 @@
+#Time complexity: O(n) where n is the number of characters in  the input string
+#Space complexity: O(n) where n is the number of characters in  the input string
```
samarthjj, line 2: Since the input can only contain lowercase or uppercase alphabets, the space should be O(52) or O(1)

samarthjj:
Good work Khushboo, please review the inline comments!

samarthjj:
Good work Khushboo, please review the inline comments!

PR #225, opened 2020-07-19
completed. please Review

satishsolanki1990:


3_longestPalindrome.py
```diff
@@ -0,0 +1,27 @@
+'''
+# time complexity = O(n)
+# space complexity = O(1)
+# working on Leetcode = yes
+# need to return max length
+'''
+
```
samarthjj, line 7: Please add your approach!

2_contiguous.py
```diff
@@ -0,0 +1,27 @@
+'''
+# time complexity = O(n)
+# space complexity = O(n)
+# working on Leetcode = yes
+# need to return max length
+'''
+
```
samarthjj, line 7: Please add your approach!

samarthjj:
Good work Satish, Please review the inline comments!
Please add your approach for every problem and solution you provide, Please use less pseudocode and more of english sentences when stating your approach!

samarthjj:
Good work Satish, Please review the inline comments!
Please add your approach for every problem and solution you provide, Please use less pseudocode and more of english sentences when stating your approach!

PR #231, opened 2020-07-28
Hashing-2

SantoshReddyKallala:


samarthjj:
Good work, Only completing the problems isn't the motive of submitting Pull Requests. Please add your approach and for the analysis use the template from **sample.java** file.

samarthjj:
Good work, Only completing the problems isn't the motive of submitting Pull Requests. Please add your approach and for the analysis use the template from **sample.java** file.

PR #265, opened 2020-08-27
Hashing-2

techmoinak:


Gupta-Keshav-001378461:
Good Work!! You don't have to work with greedy approach you can easily use hashmap for the longest palindrome problem.

techmoinak:
Hi Keshav,
Thanks for commenting. I solved the same using Hashset again after going
through today's lecture.

Thanks and Regards,
Moinak

PR #272, opened 2020-08-28
Completed Day6 Problems: Please review

ShaivyT:


Problem3.java
```diff
@@ -0,0 +1,33 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n)
```
samarthjj, line 2: O(52) or O(1), the input will only have Upper and lower case letters.

samarthjj:
Good work Shaivy, Please review the inline comments.
Please complete the **Contiguous Array** Problem as well.

samarthjj:
Good work Shaivy, Please review the inline comments.
Please complete the **Contiguous Array** Problem as well.

ShaivyT:
Please check now. Sorry I forgot to push Problem 2.

Problem3.java
ShaivyT: Can you please explain? I mean is it constant?

PR #273, opened 2020-08-28
Please review the submission

sidmal11:


LongestPalindrome.java
```diff
@@ -0,0 +1,37 @@
+// Time Complexity : O(n) since we have to traverse the entire string length
+// Space Complexity :O(n) where n is the string length , and all the characters are just once
```
samarthjj, line 2: O(52) or O(1), The input will only have upper and lower case letters.

samarthjj:
Good work Siddharth, Please review the inline comments.
Please add a brief description of your approach as well.

samarthjj:
Good work Siddharth, Please review the inline comments.
Please add a brief description of your approach as well.

PR #345, opened 2020-10-29
hashing2-submission

sujana99:


hashing2.py
```diff
@@ -0,0 +1,42 @@
+# SC -O(1)
+# TC -O(1)
```
samarthjj, line 2: TC = O(n)

hashing2.py
```diff
@@ -0,0 +1,42 @@
...
+            else:
+                values.add(i)
+        if len(values)>0:       
+            count =count+1
+        return count
+		
+		
+		
+# SC -O(1)
+# TC -O(n)
```
samarthjj, line 22: Both Time & Space here are O(n)

hashing2.py
```diff
@@ -0,0 +1,42 @@
...
+        rs=0
+        hashmap={0:-1}
+
+        for index in range(0,len(nums)):
+            if nums[index] ==0:
+                rs=rs-1
+            else:
+                rs=rs+1
+
+            if rs in hashmap.keys():
```
samarthjj, line 38: Don't use .keys(), this creates a list of all keys and performs a linear search, compromising Time.
Instead use the standard lookup of a hashmap "if rs in hashmap"

samarthjj:
Sujana, Please review the inline comments!

samarthjj:
Sujana, Please review the inline comments!

PR #379, opened 2020-12-30
Solved 2 problems

deepashree5594:


Himanshu-Agarwal93 (commented):
For problem 2, try using a hashmap to do better on the time complexity.

Himanshu-Agarwal93:
For problem 2, try using a hashmap to do better on the time complexity.

PR #390, opened 2020-12-30
Completed problem 2 and 3

shamali21:


ramya-sruthi-d (commented):
Good work, the sc for second problem is O(n) cause you are using an Arraylist.

ramya-sruthi-d:
Good work, the sc for second problem is O(n) cause you are using an Arraylist.

PR #433, opened 2021-01-26
Assignment on Hashing2_Q.1_Create SubArray_sum_equals_k, Please review

srivastavavertika201:


SubArray_sum_equals_k
```diff
@@ -0,0 +1,19 @@
+class Solution {
+    public int subarraySum(int[] nums, int k) {
+        int count =0;
```
yahska-ednihs, line 3: Please try the solution in class. You will find that there is a much better solution for this problem. Also please provide all the required information requested including the explanation of your approach

SubArray_sum_equals_k
```diff
@@ -1,3 +1,35 @@
...
+        int currentSum = 0;
```
yahska-ednihs, line 4: Good Job! Please add your approach related information in the requested format.

Contiguous array
```diff
@@ -0,0 +1,29 @@
+class Solution {
+    public int findMaxLength(int[] nums) {
```
yahska-ednihs, line 2: Good Job! Please add runtime related and other requested information.

LongestPalindrome
```diff
@@ -0,0 +1,23 @@
+class Solution {
+    public int longestPalindrome(String s) {
+        int size=0;
+        HashSet<Character> hs = new HashSet<Character>();
```
yahska-ednihs, line 4: Good Job! Please enter all the requested information including your approach.

PR #465, opened 2021-02-23
Problem1

SourabhSanghi:
Work in progress ,does not meet all test cases


Problem1
```diff
@@ -0,0 +1,38 @@
+class Solution {
+    public int subarraySum(int[] nums, int k) {
```
yahska-ednihs, line 2: Good Try. Please refer the class video for the optimal solution or reach out to the TAs or the professor.

PR #481, opened 2021-02-24
Completed two problems

divyaar0311:


kannan-mrugapphan (commented):
Good Job. try to optimize both solution using some hashing technique

kannan-mrugapphan:
Good Job. try to optimize both solution using some hashing technique

PR #620, opened 2021-06-30
Hashing-2 solution

broy2808:


Stnagam:
Please add LC link, expalin ur code with comments

Stnagam:
Please add LC link, expalin ur code with comments

PR #709, opened 2021-08-31
All three problems completed

ami-doshi:


maithrreye:
Good Job Ami ! Keep continuing the good work !

maithrreye:
Good Job Ami ! Keep continuing the good work !

PR #725, opened 2021-09-28
Hashing-2 completed, please review

srinathsai:
failing at some edge cases.

### DIFF
--- a/contigious array.java
+++ b/contigious array.java
@@ -1,0 +1,44 @@
+// Time Complexity :o(n).
+// Space Complexity :0(n).
+// Did this code successfully run on Leetcode :failed at egde cases.
+// Any problem you faced while coding this :yes, at edge casess. or when array length is 3.
+
+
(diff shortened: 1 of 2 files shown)


contigious array.java
```diff
@@ -0,0 +1,44 @@
+// Time Complexity :o(n).
```
manojvarmarudraraju, line 1: try using only one pointer to avoid confusion, consider 0 as -1 and calculate running sum

no of sub arrays
```diff
@@ -0,0 +1,47 @@
+// Time Complexity : o(n);
```
manojvarmarudraraju, line 1: For any contiguous sub-array problem you have to explore three concepts
1. Running sum
2. Two pointers
3. Sliding window
when It comes to sum of subarray running sum is the best option.

manojvarmarudraraju (commented):
Good job, please try the suggestions

manojvarmarudraraju:
Good job, please try the suggestions

PR #746, opened 2021-10-15
Create problem3.py

rucha2997:


Stnagam:
Please add comments, TC SC to all the problems!

PR #1047, opened 2022-06-02
Hashing-2 Completed in Python

tathyapatel:


LongestPelindrome.py
```diff
@@ -0,0 +1,36 @@
+#409. Longest Palindrome
+"""
+// Time Complexity : O(n) n = size of string
+// Space Complexity : O(n) n = size of hash set for worst case
```
daredevilteja, line 4: O(1) since there is a  finite number which will be stored in Hashset.

daredevilteja (changes requested):
Good work! Keep it up

daredevilteja:
Good work! Keep it up

PR #1071, opened 2022-06-28
Hashing Two 

SreekarManchukonda:


niharikagupta2021 (commented):
Good Job! Please add sc and tc as well

PR #1209, opened 2022-09-15
Completed Hashing 2 - Pls Review

lakshaynanda:


palindrome.java
```diff
@@ -1,5 +1,7 @@
 // TC: O(n)
 // SC: O(n)
```
daredevilteja, line 2: SC is O(1)

daredevilteja:
Good Work! Keep it up

palindrome.java
lakshaynanda: Yes, sorry. I'll fix this.

PR #1443, opened 2023-07-05
Completed Hashing-2

PurnaSahithiAdduri:
Completed Hashing-2

### DIFF
--- a/Problem3.java
+++ b/Problem3.java
@@ -1,0 +1,31 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode :yes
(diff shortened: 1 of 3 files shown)


Problem3.java
```diff
@@ -0,0 +1,31 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n)
```
ishika2741, line 2: Space will be O(1) , in the worst case max characters in set can be valid ascii character i.e. 256 which is constant

ishika2741 (commented):
Well done! Keep it up.

ishika2741:
Well done! Keep it up.

PR #1496, opened 2023-09-07
Completed Hashing-2. Please review

krishnadheerajkrovi:


Problem3.py
```diff
@@ -2,6 +2,9 @@
 1. Maintain a dictionary that keeps the frequency of occurances of each character in the string.
 2. If we find any character that has occurred even number of times we can definitely form a palindrome from these characters so append to a length variable.  When it is an odd occurring character, we have two cases.
 3. If it occurs once, we can have it as the middle element else we can have pairs on either side and one in middle. Handle the odd occurances with a flag to add the middle element in length and count-1 for the pairs. 
+
+TC: O(n)
+SC: O(n)
```
ishika2741, line 7: SC would be O(1) because at max we can have all lower and upper case character sin map. Which will be O(52)

ishika2741 (commented):
Well done! Keep it up.

ishika2741:
Well done! Keep it up.

PR #1707, opened 2024-06-26
Completed the hashing-2 problem set using C++

github-practice-pw:


longest_palindrome.cpp
```diff
@@ -0,0 +1,40 @@
+/*
+ *                  Longest Palindrome
+ *
+ * Leetcode : https://leetcode.com/problems/longest-palindrome/ 
+ * 
+ * Time complexity : O(N)
+ * Space complexity : O(N)
```
ishika2741: Check space complexity. Since you just have lower case characters , so it will be O(26) will be O(1)

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1742, opened 2024-07-31
Hashing-2 completed

arjunaspreethi:


LongestPalindrome.java
```diff
@@ -0,0 +1,36 @@
+//Time Complexity: O(n)
+//Space Complexity: O(n)
```
ishika2741: space used by map will be contsnat as only lower case characters are there

ishika2741 (commented):
Well done!

ishika2741:
Well done!

LongestPalindrome.java
arjunaspreethi: I have updated the space complexity here, thank you

PR #1820, opened 2025-01-02
palindrome and contiguous array

rishabh-kr-jain:


Erlybird:
Good job! What's arr in the 2nd problem?

Erlybird:
Good job! What's arr in the 2nd problem?

PR #1887, opened 2025-01-29
submit Hashing-2

shreyarajwade13:


Erlybird:
There's a syntactical error in the longest Palindrome solution. Please mention the whole solution following OOD, mention the class and write the function inside it. Just like the way mentioned in the leetcode. It's easier for us to evaluate , not all the evaluators would be from Python background, so it would be easy for everyone. Thanks.

Erlybird:
There's a syntactical error in the longest Palindrome solution. Please mention the whole solution following OOD, mention the class and write the function inside it. Just like the way mentioned in the leetcode. It's easier for us to evaluate , not all the evaluators would be from Python background, so it would be easy for everyone. Thanks.

PR #1973, opened 2025-05-22
Hashing-2 solutions in python

kedarbhusari:


dalvirushikesh:
Great work overall — clean logic! Just fix the +1 in palindrome (only add if unmatched chars exist) and remove abs() in subarray sum (must check exact diff).

PR #2007, opened 2025-05-29
Done with Hashing-2

WoodComb:


Nandha2Tech (commented):
Great Work!

Nandha2Tech:
Great Work!

super30admin:
- Strengths:
  - Correct implementation of all three problems with optimal time and space complexity.
  - Good use of comments to explain the approach.
  - Proper use of hash maps to solve the problems efficiently.
- Areas for Improvement:
  - In Problem2, the logic for handling 1s and 0s can be combined to reduce code duplication. For example, you can use `running_sum += 1 if num == 1 else -1` and then handle the hash map update and distance calculation in a single block.
  - The comments in Problem2 and Problem3 could be more detailed to explain the approach more clearly, similar to Problem1.

PR #2008, opened 2025-05-30
Hashing-2 complete

Ravi-dsa-cli:


super30admin:
- The student's implementation of `SubarraySumEqualsK.java` is excellent. It correctly uses the hashmap to track cumulative sums and efficiently counts the subarrays that sum to `k`.
- The code is clean, with good variable naming and proper use of Java collections.
- The inclusion of a `main` method for testing is a good practice, as it demonstrates the functionality with sample inputs.
- The student has shown a good understanding of the problem and the optimal approach to solve it.

PR #2016, opened 2025-06-24
Hashing 2 done

GuptaDivij:


super30admin:
- Strengths:
  - The student demonstrates a strong understanding of using HashMaps for tracking cumulative sums and counts, which is crucial for these problems.
  - The solutions are efficient and correctly handle edge cases.

- Areas for Improvement:
  - In problem1.java, the initial check `if(currSum==k) res++;` could be merged into the logic with `sumsFreq.containsKey(currSum-k)` by initializing the HashMap with `(0, 1)` to handle the case where the cumulative sum itself equals k. This would make the code slightly more concise.
  - In problem3.java, the frequency array size could be reduced to 58 (to cover all ASCII letters) or dynamically determined, but the current approach is still efficient and correct.

PR #2019, opened 2025-06-26
Completed 2 problems in Hashing-2

rachanamurty:


super30admin:
- Strengths:
  - The student demonstrates a good understanding of the problem and the use of hashing to solve it efficiently.
  - The time and space complexity analysis is accurate.

- Areas for Improvement:
  - The edge case checks in `subarray_sum_equals_k.java` (e.g., `nums.length == 1 && nums[0] == k`) are redundant and can be removed, as the main logic already covers these scenarios.
  - The comment in `contiguous_array.java` mentions "Time complexity - O(1)", which is incorrect. It should be O(n), as stated in the explanation below. This inconsistency should be corrected.

PR #2026, opened 2025-06-27
Hashing-2 Done

JPolishetty:


super30admin:
- Strengths:
  - The solutions are correct and efficiently handle the problem requirements.
  - The use of hashmap to track running sums and counts is well-implemented.
  - The code is readable and well-commented, making it easy to understand the logic.

- Areas for Improvement:
  - In Problem2, the `rSumArray` is unnecessary and can be removed to save space. The running sum can be tracked and used directly in the loop without storing it in an array.
  - In Problem3, the logic for handling odd counts could be simplified. Instead of checking `if(value != 1)`, you can always add `value - 1` for odd counts and then add 1 if there's any odd count, which is more concise.

PR #2034, opened 2025-06-29
Completed Hashing-2

shantanujoshi25:


super30admin:
- Strengths:
  - Correct implementation of all three problems with optimal time and space complexity.
  - Clear and readable code with descriptive variable names.
  - Good use of hash maps to solve the problems efficiently.

- Areas for Improvement:
  - In Problem_1.py, the condition `if( (sumArr - k) in hashMap)` could be simplified to `if sumArr - k in hashMap` for better readability.
  - In Problem_2.py, the initial condition `if(len(nums) == 1): return 0` is correct but could be omitted since the loop will handle it naturally.
  - In Problem_3.py, the final condition `if(len(s) == lenPalin): return lenPalin` is correct but could be simplified to `return lenPalin + 1 if len(s) > lenPalin else lenPalin` for brevity.

PR #2036, opened 2025-06-30
Hashing - 2

mdashora:


super30admin:
- Strengths:
  - The student has correctly implemented the running sum pattern for both problems.
  - The student has handled edge cases appropriately, such as initializing the HashMap with `(0,1)` for `SubArray.java` and `(0,-1)` for `ContiguosArray.java`.
- Areas for Improvement:
  - In `ContiguosArray.java`, the base case condition `if(nums==null && nums.length==0)` should be `if(nums==null || nums.length==0)` to handle cases where either condition is true.
  - The comments could be slightly more detailed to explain why certain initial values are added to the HashMap (e.g., why `(0,1)` or `(0,-1)` are used).

super30admin:
Strengths:
- The student demonstrates a good understanding of the running sum pattern and its application to solve subarray problems.
- The student correctly handles edge cases and initializes the hashmap appropriately.

Areas for Improvement:
- The `ContiguosArray.java` solution is for a different problem, which might be confusing if the task was specifically to solve the subarray sum problem. Ensure that the solution matches the problem statement.
- The base case check in `ContiguosArray.java` has a logical error (`nums==null && nums.length==0` should be `nums==null || nums.length==0`), though it doesn't affect the correctness in this context since the problem likely guarantees non-null input.

PR #2037, opened 2025-07-01
Hashing-2

joshiatharva1:


super30admin:
- **Correctness**: The logic for counting subarrays with sum `k` is incorrect. You should first check if `rSum - k` exists in the map before updating the map with the current `rSum`. This ensures that you are counting the correct subarrays.
- **Code Quality**: The code is readable, but you should avoid including unrelated files in the submission. Focus on the problem at hand. Also, ensure consistent variable naming and add more descriptive comments where necessary.
- **Efficiency**: The current implementation is inefficient due to the incorrect logic. Follow the reference solution's approach to ensure correctness and efficiency.

super30admin:
- **Strengths**: The student has a good understanding of the problem and has correctly identified the use of a hashmap to store running sums. The code is well-commented and follows good practices.
- **Areas for Improvement**: 
  - In `subarraySum`, the student should check for `rSum - k` before updating the map with the current `rSum`. This ensures that the count is accurate and avoids counting the current element as part of the subarray.
  - The student should also ensure that the map is updated using `map.getOrDefault(rSum, 0) + 1` to handle cases where the key is not present in the map, which is more concise and avoids potential null pointer exceptions.
  - The edge case handling for `nums.length == 1` is unnecessary if the general case is handled correctly, as the reference solution demonstrates.

PR #2050, opened 2025-07-24
all problems added

Tejas274:


super30admin:
1. Correctness:
   - Problem 1 (560-Subarray-Sum-Equals-K): The solution correctly implements the prefix sum approach with a hash map to count subarrays summing to k. It handles edge cases well.
   - Problem 2 (525-Contiguous-Array): The solution correctly uses a running sum approach to find the longest contiguous subarray with equal 0s and 1s. The initial dictionary entry {0: -1} is correctly placed.
   - Problem 3 (409-Longest-Palindrome): The solution correctly counts pairs of characters and adds 1 if there's any remaining single character for the center of the palindrome.

2. Time Complexity:
   - All solutions correctly state O(n) time complexity, which is optimal for these problems.

3. Space Complexity:
   - Problem 1 and 2 correctly state O(n) space complexity due to hash map usage.
   - Problem 3 correctly states O(1) space complexity since the character set is limited (though technically O(n) in worst case, it's constant for practical purposes due to fixed character sets).

4. Code Quality:
   - Good variable naming (rsum, count, max_length, etc.)
   - Proper spacing and indentation
   - Could improve with more comments explaining the logic, especially for Problem 1 where the rsum - k check might not be immediately obvious
   - Problem 3's initial null check is unnecessary in Python as len() of None would raise TypeError before reaching the check

5. Efficiency:
   - All solutions are optimally efficient
   - No unnecessary operations or redundant checks
   - The solutions follow standard approaches for these problems

PR #2066, opened 2025-08-03
Hashing-2

Lavanyarekkala:
Completed Hashing-2

### DIFF
--- a/contiguousArray.java
+++ b/contiguousArray.java
@@ -1,0 +1,35 @@
+/** 
+ * Time complexity: O(n)
+ * Space complexity: O(n)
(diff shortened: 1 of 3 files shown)


super30admin:
- Strengths:
  - The student has correctly implemented the logic for finding subarrays with a given sum using a hashmap.
  - The use of `getOrDefault` method is a good practice to handle cases where the key might not exist in the map.
- Areas for Improvement:
  - The variable name `runningsum` could be more conventional, such as `rSum` or `runningSum`, to align with common coding practices.
  - The comment block at the top of the file could be more concise, focusing on the key points without repeating the problem description.

PR #2074, opened 2025-08-26
Hashing-2_Done

Sarthak15997:


super30admin:
- Strengths:
  - The student has correctly implemented the logic for counting subarrays with sum equal to k.
- Areas for Improvement:
  - While the solution is correct, it would be beneficial to add a brief explanation of why the hashmap is initialized with `{0:1}` (to account for subarrays starting from the beginning of the array).
  - The variable name `hmap` could be more descriptive, such as `prefix_sum_counts`.

PR #2076, opened 2025-08-27
Implemented Hashing-2

johar-inderpreet:
- Implemented LongestPalindrome
- Implemented ContiguousArray
- Implemented SubArraySumEqualsK


super30admin:
- Strengths:
  - The student has a good understanding of the problem and has implemented the solution correctly.
  - The use of a hashmap to track running sums and their frequencies is efficient and correctly implemented.
- Areas for Improvement:
  - While the comments are helpful, some of them could be more concise. For example, the explanation of the approach in `SubArraySumEqualsK.java` is quite detailed but could be simplified.
  - The student could consider adding more edge cases in the test cases to ensure robustness (e.g., empty array, array with all elements equal to k).

PR #2079, opened 2025-08-27
completed Hashing-2

aditya18m:


super30admin:
- Strengths:
  - The code is well-commented, making it easy to understand the thought process.
  - The use of unordered_map and unordered_set is appropriate and efficient.
  - The solutions handle edge cases well, such as the dummy entries in the maps for edge cases.
- Areas for Improvement:
  - In problem1.cpp, the comment "dummy entry for edge cases" could be more specific about what edge cases it handles (e.g., when the subarray starts from the beginning).
  - In problem3.cpp, the space complexity comment for the hashset solution should be O(n) instead of O(1) since the set can grow up to the size of the input string.
  - The main functions in each file are useful for testing but could be more concise or moved to a separate test file for better organization.

PR #2083, opened 2025-08-28
hashing 2 done

desaiparam:


super30admin:
- Strengths:
  - The student has correctly implemented the solutions for all three problems.
  - The time and space complexities are optimal.
  - The use of hashmap and hashset is appropriate for the problems.

- Areas for Improvement:
  - In the "Subarray Sum Equals K" solution, the variable name `sumy` could be more descriptive, such as `running_sum` or `prefix_sum`.
  - The comment for space complexity in "Subarray Sum Equals K" and "Contiguous Array" should be corrected to O(n) instead of O(1), as the hashmap can grow up to the size of the input array in the worst case.
  - The import statement `from ast import List` in "Contiguous Array.py" is incorrect. It should be `from typing import List` as used in the other files.

PR #2089, opened 2025-08-28
Done with Hashing-2

divyam2207:


super30admin:
- Problem1.py: The solution is correct and efficient, but you can optimize it further by eliminating the prefix sum array and directly using the running sum with the hashmap. This would reduce the space complexity slightly by not storing the prefix sums.
- Problem2.py: The solution is well-implemented and follows the optimal approach. Good job!
- Problem3.py: The solution is well-implemented and follows the optimal approach. Good job!

Overall, your solutions are correct and efficient. The only minor improvement would be to optimize Problem1.py by removing the prefix sum array. Keep up the good work!

PR #2092, opened 2025-09-04
Done hashing2

ramya-gopalaswamy:


super30admin:
- Strengths:
  - The use of hash maps to optimize the solutions is commendable.

- Areas for Improvement:
  - For Problem1.cpp, consider eliminating the prefix sum array to save space, as the running sum can be computed on the fly.
  - In Problem2.cpp, ensure that the conversion of 0s to -1s is clearly documented to explain the logic behind it.
  - In Problem3.cpp, the comment "use even part, leave 1 aside" could be more explicit about why this is done (to allow for a central character in the palindrome).

PR #2096, opened 2025-09-17
Solved Hashing Problems

aahiawasthee:


super30admin:
The student has done an excellent job in solving the "sub-array-sum-equals-k" problem. The solution is correct, efficient, and well-written. The comments provided are clear and help in understanding the logic. The student has also correctly identified the time and space complexity of the solution.

For the other two problems ("contiguous-array" and "longest-palindrome"), the solutions are also correct and efficient. The comments and explanations provided are helpful. However, the space complexity for "longest-palindrome" is noted as O(1), which is correct in practice due to the limited number of unique characters, but technically it is O(n) in the worst case if all characters are unique. This is a minor point and does not affect the overall correctness of the solution.

Overall, the student has demonstrated a strong understanding of the problems and has provided high-quality solutions.

PR #2104, opened 2025-10-09
Hashing-2 solutions

kalyan-ng22:


super30admin:
- The student's solution for `SubArraySumEqualsK.java` is excellent. It correctly implements the logic to find subarrays with sum equal to k using a HashMap to store running sums and their counts. The approach is efficient and matches the reference solution in both time and space complexity.
- The comments are helpful and explain the approach clearly, which is good practice for maintainability.
- One minor suggestion would be to consider using a for-each loop for iterating over the array, as it can make the code slightly cleaner, but this is a minor stylistic preference and does not affect correctness or performance.

PR #2105, opened 2025-10-10
Hashing-2

bharath6969:


super30admin:
Strengths:
- The solutions are correct and efficient, with optimal time and space complexity.
- The logic is clear and well-implemented.
- The use of hash maps/sets is appropriate for the problems.

Areas for Improvement:
- Variable naming could be more consistent and descriptive (e.g., `intial_count` should be `initial_count`).
- The `Longestpalindorm_hashmap.py` solution is redundant and could be simplified by using a set, as done in `LongestPalindrome_hashset.py`.
- The README.md could be more concise and focused on the problem at hand (`subarraySum`), rather than including explanations for other problems.
- The comments in the code could be more detailed to explain the logic, especially for the `findMaxLength` solution.

PR #2114, opened 2025-10-25
Hashing-2

lalitha7571:
Hashing-2


super30admin:
- **Strengths**: The student has correctly implemented the solution for the `subarraySum` problem, demonstrating a good understanding of the problem and the use of hash maps to optimize the solution.
- **Areas for Improvement**: 
  - The nested `if` conditions for updating the map can be simplified using `map.getOrDefault` to make the code cleaner and slightly more efficient.
  - The student should ensure that only the relevant code for the problem being evaluated is included in the submission to avoid confusion.

PR #2115, opened 2025-11-13
Done Hashing2

Bhumika-95:


super30admin:
- The student has demonstrated a strong understanding of the problems and provided correct and efficient solutions.
- The code is clean and follows good practices, making it easy to read and understand.
- The student correctly identified and documented the time and space complexities for each solution.
- One minor suggestion: for the `LongestPalindrome` solution, the boolean array size is hardcoded to 95 (assuming ASCII values from 'A' to 'z'). It might be better to use a larger size (e.g., 128) to cover all ASCII characters or use a HashSet for more flexibility with Unicode characters.

PR #2117, opened 2025-11-19
Hashing-2 completed

Tejaswini2112:


super30admin:
Strengths:
- The student has provided both optimized and brute-force solutions, which is excellent for understanding the problem from different perspectives.
- The code is well-commented and easy to follow.
- The use of hash maps and hash sets is appropriate and efficient.

Areas for Improvement:
- The problem title "Hashing-2" is not clear. It would be better to have a more descriptive title or to include the problem description.
- The student could consider adding more edge cases in the main method to test the robustness of the solutions.
- For the LongestPalindrome problem, the optimized solution could be slightly improved by checking if the string length is odd or even at the beginning to avoid the final check for an empty set.

PR #2118, opened 2025-11-20
Done

PurvaT-11:


super30admin:
- The student's solution is well-implemented and correctly solves the problem. The use of a dictionary to track running sums and their first occurrences is a smart approach.
- The comments are helpful but could be slightly more concise. For example, the explanation of the edge case handling could be simplified.
- The code is clean and easy to follow, which is excellent. However, the student could consider adding a brief docstring at the beginning of the function to summarize its purpose and parameters, which is a good practice for maintainability.

PR #2135, opened 2025-12-31
Completed Hashing 2 problems

thondaha:


super30admin:
Strengths:
- The student has correctly implemented the solutions for all three problems, demonstrating a good understanding of the concepts.
- The use of hash maps to track running sums and frequencies is correctly applied.
- The test cases in `TestHashing2.java` cover various scenarios, showing good testing practices.

Areas for Improvement:
- In `BinarySubArray`, the logic for updating `rSum` could be more intuitive if it increments for 1 and decrements for 0 (the current implementation does the opposite). While this doesn't affect correctness, it might be more intuitive to follow the problem's natural interpretation.
- The comments in `SubArraySumK` could be slightly more detailed to explain why `map.put(0, 1)` is necessary (i.e., to account for subarrays starting from the beginning of the array).
- The test cases could include edge cases like empty arrays or strings to ensure robustness.

PR #2136, opened 2026-01-01
Complete Hashing2

JainShubham23:


super30admin:
- Strengths:
  - The student has a good understanding of the prefix sum technique and has applied it correctly in Problem1 and Problem2.
  - The use of a hashset in Problem3 to count pairs for the longest palindrome is efficient and well-implemented.
  - The code is well-commented, making it easy to understand the thought process behind each solution.
- Areas for Improvement:
  - In Problem3, the time and space complexity were not filled in the comments. It would be good to complete these details for consistency.
  - The comments in Problem1 could be slightly more concise while still being clear. For example, the explanation of the base case could be simplified.

PR #2143, opened 2026-01-06
subarraySum

dbkuppagiri:


super30admin:
- Strengths:
  - The solution correctly implements the prefix sum and hash map approach.
- Areas for Improvement:
  - Instead of storing arrays of indices in the hash map, consider storing counts of how many times each cumulative sum has occurred. This simplifies the logic and reduces memory usage.
  - The initial comment could be more detailed to explain the intuition behind the approach, especially why storing counts would be more efficient.
  - The variable name `currAssignedVal` could be more descriptive, such as `indices` or `sumIndices`.

PR #2144, opened 2026-01-07
LC 525: Contiguous Array

dbkuppagiri:


super30admin:
- The solution is well-thought-out and correctly implements the prefix sum approach with a hash map.
- The initialization of `currCount` to 1 and storing it at index -1 is a clever way to handle subarrays starting from the beginning of the array.
- The logic for updating `maxLength` when a matching count is found is correct and efficient.
- The code is clean and easy to follow, with good variable naming and clear comments explaining the intuition.
- One minor suggestion: The comment "Intution" should be corrected to "Intuition" for better readability.

PR #2147, opened 2026-01-11
Hashing-2 HW LongestPalindrome and SubarraySumEqualsK

Amaan-29:


super30admin:
1. **Correctness**: The solution correctly implements the logic to find subarrays with sum equal to k using a hashmap to store running sums and their counts. This matches the reference solution.
2. **Time Complexity**: The solution has a time complexity of O(n), which is optimal and matches the reference solution.
3. **Space Complexity**: The solution uses O(n) space, which is also optimal and matches the reference solution.
4. **Code Quality**: The code is well-structured and readable. However, the use of `map.getOrDefault(rSum, 0) + 1` in the reference solution is more concise and avoids the need for an additional `if-else` block. Consider using this approach for cleaner code.
5. **Efficiency**: The solution is efficient, but the additional `if-else` block can be simplified using `map.getOrDefault` as mentioned above. This would make the code more concise and potentially slightly more efficient.

Overall, the solution is very good, but minor improvements in code style could make it even better.

PR #2158, opened 2026-02-11
[Hashing based techiques]

takchiks:


super30admin:
Strengths:
- The solution in `SubarraySum.java` is correct and efficient, using the optimal approach with a hashmap.

Areas for Improvement:
- Avoid including multiple solutions to different problems in the same submission. Focus on the problem at hand.
- The brute-force solution, while correct, is inefficient and should not be used for this problem due to high time complexity.
- Ensure that the submission only contains the relevant code for the problem. Extraneous code can cause confusion and may lead to errors in evaluation.

PR #2159, opened 2026-02-11
Completed Hashing-2

sarvanibaru:


super30admin:
Your solution for "Subarray Sum Equals K" is excellent. Here are some strengths and minor suggestions:

Strengths:
- You have correctly implemented the optimal solution using a hash map to store cumulative sums.
- The code is well-commented, explaining the approach clearly.
- You handled the edge case by initializing the map with (0,1).
- The use of `map.getOrDefault` is efficient and concise.

Areas for Improvement:
- There are no major issues. However, you might consider using a for-each loop for iterating over `nums` to make the code slightly more concise. For example:
  ```java
  for (int num : nums) {
      rSum += num;
      // ...
  }
  ```
  But this is a minor stylistic point and does not affect correctness.

- The variable name `cmp` might be slightly ambiguous. Consider renaming it to something like `complement` or `target` to make it more descriptive. However, it is understandable in context.

Overall, your solution is correct and efficient.

PR #2161, opened 2026-02-15
S30 FAANMG Problem #19 | Longest Palindrome

sandy7907:


super30admin:
It seems there might have been a mix-up in the problem you were solving. The problem you provided code for is "Longest Palindrome", but the problem you were asked to solve is "Subarray Sum Equals K". 

For the "Subarray Sum Equals K" problem, you need to count the number of contiguous subarrays that sum to k. The common efficient solution uses a hash map to store cumulative sums and their frequencies. 

Here is a brief explanation of the correct approach:
- Initialize a hash map to store cumulative sums (prefix sums) and their counts.
- Start with cumulative sum 0 having a count of 1.
- Iterate through the array, updating the cumulative sum.
- For each cumulative sum, check if (cumulative sum - k) exists in the map. If it does, add its count to the result.
- Update the map with the current cumulative sum.

Your code for the longest palindrome is correct for that problem, but it does not address the current problem. Please ensure you are solving the correct problem.

super30admin:
Strengths:
- The solution correctly implements the running sum technique with a hash map to efficiently count subarrays summing to k.
- The code includes helpful comments that explain the approach.
- The use of `merge` method for updating the map is concise and modern.

Areas for Improvement:
- The problem requires solving only "Subarray Sum Equals K", so the other files (ContiguousArray.java, LongestPalindrome.java, and README.md) are unnecessary and should be removed to avoid confusion.
- While the comments are good, they could be slightly more detailed to explain why the initial entry (0,1) is added to the map.
- The variable name `rSumCounterMap` is a bit verbose; `prefixSumCount` or similar might be more conventional.

Overall, the solution for SubArraySum.java is excellent. However, since the submission includes extraneous files, it might indicate a misunderstanding of the problem requirements. For future submissions, ensure only the relevant code file is provided.

PR #2162, opened 2026-02-15
Hashing 2 

sandy7907:


super30admin:
Your solution for SubArraySum.java is excellent. It correctly implements the running sum technique with a hash map, and your comments at the top clearly explain the approach. The code is efficient and follows best practices. However, when submitting a solution for a specific problem, ensure that you only include the relevant code file. Including unrelated files (like ContiguousArray.java, LongestPalindrome.java, and README.md) can cause confusion and may not be accepted by automated grading systems. Focus on the problem at hand and avoid submitting extra files unless explicitly required.

super30admin:
Your solution for SubArraySum.java is excellent. You correctly implemented the running sum approach with a hash map, and your code is clean and well-commented. The use of `merge` method for updating the map is concise and modern. However, note that the problem requires solving only the Subarray Sum Equals K problem, so including solutions for other problems in the same submission might be confusing. In the future, ensure that your submission only contains the code relevant to the problem at hand. Also, while your comments are helpful, make sure they are accurate and concise. For instance, the comment "Add a default entry in the map with [0, 1] to make sure if the comp(rSum -k) is 0 then the array until that element will be 1 sub array" is correct, but it could be phrased more clearly. Overall, great job!

PR #2164, opened 2026-02-26
Problem3 Longest palindrome added

megharaykar:


super30admin:
It seems there has been a mix-up in the problem you are solving. The problem you provided code for is "Longest Palindrome", but the problem you were assigned is "Subarray Sum Equals K". For the correct problem, you need to count the number of contiguous subarrays that sum to k. 

To solve "Subarray Sum Equals K", you should use a prefix sum approach with a hash map to store cumulative sums and their frequencies. Here's a brief outline of the correct approach:
- Initialize a hash map to store cumulative sums (prefix sums) and their counts.
- Start with cumulative sum 0 having a count of 1.
- Iterate through the array, updating the cumulative sum.
- For each cumulative sum, check if (cumulative sum - k) exists in the hash map. If it does, add its count to the result.
- Update the hash map with the current cumulative sum.

Your current code for "Longest Palindrome" is well-structured and follows a logical approach for that problem, but it is not relevant here. Please ensure you are working on the correct problem statement.

PR #2169, opened 2026-04-16
hashing-2

spencerkrebs:


super30admin:
### Subarray Sum Equals K (subarray-sum-equals-k.py)
Your approach is on the right track and demonstrates a good understanding of the prefix sum technique. However, there is a critical issue in the order of operations. You should update the hash map after checking for `curSum - k` but before the next iteration. Also, note that the check for `curSum == k` is redundant because when `curSum == k`, `curSum - k` is 0, and if you have stored 0 in the hash map with count 1 initially, it will be captured in the `h[curSum - k]` check. Therefore, you can remove the explicit check for `curSum == k` and rely on the hash map check. 

To fix the solution, you should:
1. Initialize the hash map with `h[0] = 1` to account for the prefix sum 0 at the start.
2. Remove the explicit check for `curSum == k` because it is already handled by `h[curSum - k]` when `curSum - k == 0`.
3. Update the hash map after the count increment to avoid counting the current prefix sum itself when checking for `curSum - k`.

Here is the corrected version of your code:
```python
class Solution:
    def subarraySum(self, nums: List[int], k: int) -> int:
        curSum = 0
        count = 0
        h = defaultdict(int)
        h[0] = 1  # Initialize with prefix sum 0
        for num in nums:
            curSum += num
            # Check how many times we've seen (curSum - k)
            count += h[curSum - k]
            # Then update the hash map for the current prefix sum
            h[curSum] += 1
        return count
```

This change ensures that when we check for `h[curSum - k]`, we are only considering prefix sums that occurred before the current index, which is necessary to avoid counting the current element as part of the subarray.

VERDICT: NEEDS_IMPROVEMENT

---

### Contiguous Binary Subarray (contiguous-array.py)
Your approach of tracking the difference between ones and zeros is on the right track, which is similar to the reference solution that uses a running sum (where 0 is treated as -1 and 1 as +1). However, there are several issues:

1. Initialization: You need to initialize the diff_index map with a key of 0 at index -1. This is important because when the running sum (or difference) is 0, it means we have equal zeros and ones from the start up to the current index. Without this, you might miss subarrays that start from the beginning.

2. Updating the result: Your condition for when one == zero is correct, but you set res to one+zero, which is actually the total length from start to current (which is i+1). However, this should be handled by the same mechanism as other cases. In the reference solution, when the running sum is 0, it calculates the length as i - (-1) = i+1.

3. The else branch: You are trying to use the stored index for the current difference (one-zero) to calculate the length. However, the correct approach is to check if the current running sum (or difference) has been seen before. If it has, then the subarray from the previous index+1 to the current index has equal zeros and ones. So you should store the first occurrence of each running sum (or difference) and then when you see the same value again, you calculate the length between the current index and the stored index.

4. The running sum: In your code, you are using one - zero as the key. This is equivalent to the running sum if we treat 0 as -1 and 1 as +1. However, note that the reference solution uses a running sum that starts at 0 and then decrements for 0 and increments for 1. Your one - zero is exactly the same as the running sum in the reference solution. So that part is correct.

5. Storing the index: You should only store the index for a running sum if it hasn't been seen before. This ensures that you have the earliest occurrence, which gives the longest subarray for that running sum.

6. The res update: In the else branch, you are always updating res with i - idx, but this should only be done if the running sum is already in the map. Also, you are not considering the case when the running sum is 0 (which you handle separately) but actually it can be handled uniformly.

Suggested correction:
- Initialize diff_index with {0: -1} to account for the starting point.
- For each index, calculate the running sum (one - zero) or equivalently use a variable (say, rSum) that starts at 0 and updates by +1 for 1 and -1 for 0.
- Check if the current rSum is in diff_index. If it is, then update res = max(res, i - diff_index[rSum]). Otherwise, add rSum to diff_index with value i.
- Remove the separate condition for one == zero because it is already handled by the above.

Here's how your code should look after correction:
```python
class Solution:
    def findMaxLength(self, nums: List[int]) -> int:
        rSum = 0
        res = 0
        diff_index = {0: -1}  # initialize with 0 at index -1
        
        for i, n in enumerate(nums):
            if n == 0:
                rSum -= 1
            else:
                rSum += 1
                
            if rSum in diff_index:
                res = max(res, i - diff_index[rSum])
            else:
                diff_index[rSum] = i
                
        return res
```

This corrected version uses the same logic as the reference solution and should work correctly.

VERDICT: NEEDS_IMPROVEMENT

---

### Longest Palindrome in a string (longest-palindrome.py)
Remember, the student's solution is written in Python, while the reference is in Java. The logic should be similar, but language-specific differences are acceptable.

Let's begin.

VERDICT: NEEDS_IMPROVEMENT

PR #2171, opened 2026-04-16
Hashing-2

hiteshmadapathi:


super30admin:
### Subarray Sum Equals K (Problem1.py)
Your solution is on the right track and correctly applies the running sum technique with a hashmap. This is an efficient approach for solving the problem. Here are some points for improvement:

1. **Indentation**: Python requires consistent indentation. The code inside the for-loop should be indented properly. For example:
   ```python
   for i in range(len(nums)):
        rsum = rsum + nums[i]
        # ... rest of the code inside the loop
   ```

2. **Variable Names**: Consider using more descriptive variable names. For instance, `rsum` is clear, but `op` might be better named as `count` or `result` to make the code more readable.

3. **Efficiency in Hashmap Update**: Instead of using an if-else block to update the hashmap, you can use `hmap.get(rsum, 0) + 1` to simplify. This is similar to the Java reference solution's `getOrDefault`. For example:
   ```python
   hmap[rsum] = hmap.get(rsum, 0) + 1
   ```
   This reduces the code lines and improves readability.

4. **Looping Directly Over Elements**: Since you don't use the index `i` for anything other than accessing `nums[i]`, you can iterate directly over the elements in `nums` to make the code cleaner:
   ```python
   for num in nums:
        rsum += num
        # ...
   ```

5. **Initialization**: The initialization of `hmap` with `{0:1}` is correct and necessary.

Here's a revised version of your code with these improvements:
```python
class Solution:
    def subarraySum(self, nums: List[int], k: int) -> int:
        rsum = 0
        hmap = {0: 1}
        count = 0

        for num in nums:
            rsum += num
            if rsum - k in hmap:
                count += hmap[rsum - k]
            hmap[rsum] = hmap.get(rsum, 0) + 1
        
        return count
```

Overall, your understanding of the algorithm is good, but pay attention to Python-specific details like indentation and code style.

VERDICT: PASS

---

### Contiguous Binary Subarray (Problem2.py)
Strengths:
- The code is concise and follows the same logic as the reference solution.
- The use of a hash map to store the first occurrence of each running sum is correct and efficient.
- The initial setup with `hmap[0] = -1` is crucial for handling subarrays starting from index 0, and it is correctly implemented.
- The variable names are clear and the code is readable.

Areas for Improvement:
- While the code is correct, it could be slightly optimized by avoiding the repeated `max` function call inside the loop. Instead, you can update the maximum only when the running sum is found in the map. This is already done correctly, but note that the `max` function is called only when the running sum is found, which is efficient.
- The comment mentions "first occurence and last occurence", but actually we are storing the first occurrence of each running sum. The term "last occurence" might be misleading; it's better to say that when the same running sum appears again, the subarray between the first occurrence and the current index has equal 0s and 1s.
- The code does not handle the case where the input array is empty, but the problem does not specify that the array is non-empty. However, since the initial map has an entry for 0 at index -1, and the loop runs from 0 to len(nums)-1, it should return 0 for an empty array, which is correct because there are no elements.

Overall, the solution is well-written and efficient.

VERDICT: PASS

---

### Longest Palindrome in a string (Problem3.py)
Your solution is correct and efficient, matching the expected time and space complexity. Well done! 

To improve code quality:
1. Consider using more descriptive variable names. For example, `op` could be renamed to `count` or `pair_count` to make the code clearer.
2. Instead of iterating by index, you can directly iterate over the characters in the string: `for char in s`. This is more Pythonic and avoids unnecessary indexing.
3. You can also combine the condition for adding an extra character at the end more concisely by checking if the set is non-empty.

Here's a slightly revised version of your code with these improvements:
```python
class Solution:
    def longestPalindrome(self, s: str) -> int:
        char_set = set()
        count = 0
        for char in s:
            if char in char_set:
                count += 2
                char_set.remove(char)
            else:
                char_set.add(char)
        if char_set:
            count += 1
        return count
```

These changes enhance readability without altering the logic. Keep up the good work!

VERDICT: PASS

PR #2174, opened 2026-04-19
Hashing 2

MeghaN28:


super30admin:
### Subarray Sum Equals K (SubArraySumK.java)
Strengths:
- The code is clean and follows good coding practices.
- The comments are helpful and explain the approach clearly.
- The solution efficiently computes the result in a single pass through the array.

Areas for Improvement:
- The variable name "compt" might be better named as "complement" or "target" for clarity, as "compt" is not immediately intuitive.
- The comment "always update prefix sum frequency" is correct, but it might be better placed after the update operation to avoid confusion. However, the code logic is correct.

Overall, this is an excellent solution that demonstrates a strong understanding of the prefix sum technique.

VERDICT: PASS

---

### Contiguous Binary Subarray (LongestPalindrome.java)
It appears that you have submitted a solution for the "Longest Palindrome" problem instead of the "Contiguous Binary Subarray" problem. Please double-check the problem you are solving and ensure you are uploading the correct file. For the "Contiguous Binary Subarray" problem, you need to use a different approach, such as using a hash map to track the running sum (where you treat 0 as -1 and 1 as +1) and look for indices where the running sum is the same, indicating an equal number of 0s and 1s in between. I recommend revisiting the problem statement and implementing the correct solution.

VERDICT: NEEDS_IMPROVEMENT

---

### Longest Palindrome in a string (Solution.java)
It seems there has been a mix-up in the problem you are solving. The problem you are asked to solve is about finding the length of the longest palindrome that can be built from the characters in the string. Your current solution is for a different problem (Contiguous Array). 

For the correct problem, the key idea is to count the frequency of each character. A palindrome can have at most one character with an odd frequency (which would be the center). The rest of the characters must have even frequencies. You can use a HashSet to keep track of pairs: when you see a character for the second time, you can remove it from the set and add 2 to the count. If there are any characters left in the set at the end, you can add 1 for the center.

Here is a step-by-step approach for the correct problem:
1. Initialize a HashSet to keep track of characters that appear an odd number of times.
2. Initialize a count variable to 0.
3. Iterate over each character in the string:
   - If the character is already in the set, remove it and add 2 to the count (because you found a pair).
   - Otherwise, add the character to the set.
4. After processing all characters, if the set is not empty, you can use one character as the center, so add 1 to the count.
5. Return the count.

This approach has O(n) time complexity and O(1) space complexity (since the number of distinct characters is bounded).

You should also note that the problem states that letters are case-sensitive, so 'A' and 'a' are considered different.

VERDICT: NEEDS_IMPROVEMENT

PR #2177, opened 2026-05-04
Solved Problem 1 - Subarray sum equals k, Problem 2  -Contiguous array and Problem 3 - Longest palindrome

anirudhv98:
Completed Hashing-2, Please review

### DIFF
--- a/Problem1.cs
+++ b/Problem1.cs
@@ -1,0 +1,45 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No
+
+// Your code here along with comments explaining your approach
+/*
+At first, I put the key-value pair 0,1 in the dictionary as running sum 0 exists in the dictionary once. I iterate through each
+element calculating the running sum. If running sum is not present in the dictionary, I put the running sum as key and count of how
+many times it has appeared so far as 1. Else I update the count for running sum. I check if dictionary contains rSum - k (We know the 
+running sum, we know the value k. So rSum - k would give us the values of elements that sum up to k)
+*/
+
+
+public class Solution
+{
+    public int SubarraySum(int[] nums, int k)
+    {
+        int rSum = 0, result = 0;
+        Dictionary<int, int> dict = new();
+        dict[0] = 1;
+
+        for (int i = 0; i < nums.Length; i++)
+        {
+            rSum += nums[i];
+
+            if (dict.ContainsKey(rSum - k))
+            {
+                result += dict[rSum - k];
+            }
+
+            if (!dict.ContainsKey(rSum))
+            {
+                dict[rSum] = 1;
+            }
+
+            else
+            {
+                dict[rSum] += 1;
+            }
+        }
+
+        return result;
+    }
+}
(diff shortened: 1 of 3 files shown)


super30admin:
### Subarray Sum Equals K (Problem1.cs)
Your solution is well-implemented and follows the optimal approach for this problem. Here are some points to consider:
1. **Correctness**: Your code correctly calculates the number of subarrays by maintaining a dictionary of cumulative sums. The initial condition `dict[0] = 1` is correctly set to account for subarrays starting from index 0.
2. **Time and Space Complexity**: You have achieved O(n) for both, which is optimal for this problem.
3. **Code Quality**: The code is readable and well-commented. However, you can simplify the dictionary update using C#'s `TryGetValue` or by using a more concise method. For example, you can replace the if-else block with:
   ```csharp
   dict[rSum] = dict.GetValueOrDefault(rSum, 0) + 1;
   ```
   But note that `GetValueOrDefault` is available in .NET Core and later versions. Alternatively, you can use:
   ```csharp
   if (dict.ContainsKey(rSum)) {
        dict[rSum]++;
   } else {
        dict[rSum] = 1;
   }
   ```
   Your current approach is correct, but the above might be slightly more efficient or idiomatic. However, your code is clear as is.
4. **Efficiency**: The solution is efficient. One minor improvement: you can check for `rSum - k` after updating the dictionary? Actually, no, because you are updating the dictionary after the check, which is correct. If you updated before, you might count the current element incorrectly. So your order is correct.
5. **Edge Cases**: Your solution handles cases where k is negative or zero, and when nums contains negative numbers, because the cumulative sum approach naturally accounts for these.

Overall, this is a solid solution. Keep up the good work!

VERDICT: PASS

---

### Contiguous Binary Subarray (Problem2.cs)
Strengths:
- The solution correctly implements the algorithm using a dictionary to track running sums.
- The initialization of the dictionary with (0, -1) is correctly handled, which is crucial for calculating the length of subarrays starting from index 0.

Areas for Improvement:
- There is a minor issue in the logic: when the running sum is already present in the dictionary, the solution does not update the dictionary with the current index. This is actually correct because we want the first occurrence of the running sum to maximize the subarray length. However, the code could be slightly more explicit in its intent. For example, the condition `if (!dict.ContainsKey(rSum))` is used to store the first occurrence, and the else branch calculates the length. This is correct, but it might be helpful to add a comment explaining why we don't update the dictionary when the key exists.
- The variable name `rSum` is clear, but it could be spelled out as `runningSum` for even better readability, though this is a minor point.

Overall, the solution is excellent and follows the same approach as the reference solution.

VERDICT: PASS

---

### Longest Palindrome in a string (Problem3.cs)
Your solution is well-implemented and correctly addresses the problem. Here are some strengths and minor suggestions:

Strengths:
- Correct logic: You correctly count pairs of characters and add 1 for any remaining odd-count characters at the end.
- Efficiency: The solution has optimal time and space complexity.
- Code clarity: The code is readable and well-structured with appropriate variable names.

Areas for Improvement:
- Although the space complexity is effectively O(1) due to the fixed character set, it's good to note that in the context of the problem, the HashSet size is bounded by 52, making it constant. You might want to mention this in your comments.
- Consider using a foreach loop for cleaner iteration over the string characters, which can enhance readability. For example: `foreach (char c in s)` instead of using an index.

VERDICT: PASS

PR #2178, opened 2026-05-04
Adding Hashing-2 solutions (All 3)

nikhylw:
Problem 1 - W2_5_560_subarray_sum_equals_k.py
Problem 2 - W2_4_525_contiguous_array.py
Problem 3 - W2_3_409_longest_palindrome.py


super30admin:
### Subarray Sum Equals K (W2_5_560_subarray_sum_equals_k.py)
1. The solution correctly uses the prefix sum approach with a hashmap to count subarrays summing to k. The logic is similar to the reference solution.
2. However, the code checks if the current prefix sum (rsum) is in the hashmap before incrementing its count. This is unnecessary because the `getOrDefault` method in Java (or its equivalent in Python) can be replaced with a more efficient approach. In Python, you can use `hashmap.get(rsum, 0)` to avoid the extra check.
3. Specifically, the lines `if rsum not in hashmap: hashmap[rsum] = 0` can be replaced with a single line: `hashmap[rsum] = hashmap.get(rsum, 0) + 1`. This is more concise and avoids an extra lookup.
4. The variable naming is clear, and the code is well-structured. The comments at the top are helpful for understanding the complexity.
5. The solution correctly initializes the hashmap with {0:1} to account for subarrays starting from index 0. This is crucial for the algorithm to work.

VERDICT: PASS

---

### Contiguous Binary Subarray (W2_4_525_contiguous_array.py)
Your solution is excellent! It correctly implements the algorithm and has optimal time and space complexity. Here are a few minor suggestions for improvement:
1. Consider adding a brief comment explaining the approach, such as how the running sum and hashmap work together to find the maximum length.
2. The variable name "maxx" could be changed to "max_len" for clarity, but it's not a major issue.
3. You might consider using enumerate for better Pythonic style, but the current loop is fine.

Overall, great job! Your solution is efficient and clear.

VERDICT: PASS

---

### Longest Palindrome in a string (W2_3_409_longest_palindrome.py)
Your solution is excellent! It correctly implements the algorithm and has optimal time and space complexity. Here are a few minor suggestions to enhance code quality:

1. **Iterate directly over characters**: Instead of using `for i in range(0, len(s)):`, you can directly iterate over the characters in the string with `for ch in s:`. This is more Pythonic and avoids indexing.

2. **Use set operations efficiently**: Your use of the set is correct. However, you can use the set's methods more idiomatically. For example, you can use `if ch in hashset` as you did, but note that the `in` operator is efficient for sets.

3. **Variable naming**: The variable `hashset` is clear, but in Python, we often use simple names like `char_set` or `odd_chars` to indicate the purpose. This is a minor point.

4. **Consistency**: The code is consistent and follows good practices. The comments are helpful.

Here's a slightly revised version of your code with these suggestions:

```python
class Solution:
    def longestPalindrome(self, s: str) -> int:
        odd_chars = set()
        count = 0

        for ch in s:
            if ch in odd_chars:
                count += 2
                odd_chars.remove(ch)
            else:
                odd_chars.add(ch)
        
        if odd_chars:
            count += 1
        
        return count
```

Overall, great job! Your solution is efficient and correct.

VERDICT: PASS

PR #2179, opened 2026-05-12
Added Solutions for hashing

LeelaTotapally:


super30admin:
**Strengths:**
- Good use of a hashmap for storing cumulative sums
- Correct initialization of map with {0,1}
- Clean variable naming and reasonable formatting

**Areas for Improvement:**
1. **Critical Logic Error**: The algorithm needs to check during iteration if there exists a previous cumulative sum such that `currentSum - previousSum = k`. This check must happen inside the main loop, not in a separate loop over map keys.

2. **Algorithm Structure**: The correct approach is:
   - For each element, update running sum
   - Check if `runningSum - k` exists in map (meaning subarray ending here sums to k)
   - Add the frequency of that sum to result
   - Then increment the frequency of current running sum

3. **Time Complexity Issue**: The current solution has O(n) for first loop but O(m) for second loop where m is the number of unique cumulative sums. While still O(n) overall, the logic is fundamentally broken.

4. **Simplify the Code**: The if-else block for updating the map can be simplified to a single `map.put(rSum, map.getOrDefault(rSum, 0) + 1)` call.

**Recommended Fix**: Move the check for subarray sums inside the main loop and use the correct comparison `map.containsKey(rSum - k)`.

VERDICT: NEEDS_IMPROVEMENT

---

**Strengths:**
- Correctly implements the prefix sum technique with HashMap
- Clear and informative comments explaining the approach
- Good variable naming conventions
- Properly handles the base case with map.put(0, -1)

**Areas for Improvement:**
- The code could be slightly more concise by combining the if-else branches for updating rSum into a single line using a ternary operator: `rSum += (nums[i] == 1) ? 1 : -1;`
- Consider using `var` keyword in Java 10+ for cleaner code, though this is optional

**Minor Note:**
The solution is functionally equivalent to the reference solution and achieves the same time and space complexity. The implementation is correct and follows good coding practices.

VERDICT: PASS

---

**Strengths:**
- The solution correctly solves the problem with proper logic
- Good use of comments explaining the approach
- Clear variable naming
- Proper handling of both even and odd character counts

**Areas for Improvement:**
- Space complexity could be improved from O(n) to O(1) by using a HashSet instead of HashMap (since we only need to track which characters appear an odd number of times)
- The character counting logic is verbose - consider using `map.getOrDefault(c[i], 0) + 1` or `map.merge()` for cleaner code
- The `temp` variable name doesn't clearly convey its purpose as an "odd count exists" flag

VERDICT: PASS

PR #2189, opened 2026-06-15
hashing-2

joshyoon92:


super30admin:
### Subarray Sum Equals K (subarraySumEqual.java)
**Strengths:**
- Good understanding of the prefix sum + hashmap approach for this problem
- Correct algorithm logic (tracking running sum and checking for `runningSum - k`)
- Proper initialization of map with `{0: 1}` to handle subarrays starting from index 0

**Areas for Improvement:**
1. **Critical Bug:** Fix the variable name inconsistency - you're using `arr` as parameter but `nums` inside the loop. Use consistent naming.
2. **Critical Bug:** Change return type from `boolean` to `int` since the problem requires returning the count.
3. **Code Style:** Use `map.getOrDefault(runningSum, 0) + 1` instead of the if-else block for cleaner code.
4. **Naming:** Use descriptive names that match the problem (`nums` and `k` are standard in LeetCode).
5. **Class Structure:** The class name `SubarraySumEqual` should follow Java naming conventions (PascalCase is correct, but could be more descriptive like `SubarraySumEqualsK`).

VERDICT: NEEDS_IMPROVEMENT

---

**Strengths:**
- Good understanding of the prefix sum/hashmap approach for this problem
- Correct overall algorithm structure
- Clean variable naming (rSum, max)

**Areas for Improvement:**
1. **Critical Bug**: Move `hm.put(0,-1);` outside and before the for loop. Currently it's inside the loop, causing it to execute on every iteration and overwriting the initial condition needed for subarrays starting at index 0.
2. **Logic Flow**: The current structure has the initial map entry being set on every iteration, which is inefficient and causes the bug. The initial entry should only be set once before iterating.
3. **Consider using `computeIfAbsent` or `getOrDefault`** for cleaner code, though the current approach can work with the fix.

The core algorithm idea is correct - using a running sum where 0 decrements and 1 increments, then finding when the same sum repeats to calculate subarray length. With the fix to move the initial map entry outside the loop, the solution should work correctly.

VERDICT: NEEDS_IMPROVEMENT

---

### Longest Palindrome in a string (LongestPalin.java)
Strengths:
- Correct algorithm implementation that matches the reference solution
- Good use of HashSet for tracking unpaired characters
- Proper handling of the odd-count character case

Areas for Improvement:
- Consider using enhanced for-each loop: `for (char c : s.toCharArray())` instead of index-based iteration for cleaner code
- Variable naming could be more descriptive (e.g., `unpairedChars` instead of `hs`)
- The `contains()` check before `add()` is redundant; you can use the return value of `add()` directly

VERDICT: PASS

